PR #5, opened 2021-01-07
NF-43 - Restrict access to front-end service to logged in GG user

SteveSugden:
Applied an 'IdentifierAction' filter to the default end-point (was /hello-world) which ensures that the user is logged in via GG
This approach was taken from NDRC which in turn was generated using the Jenkins scafford job.

### DIFF
--- a/app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/views/SessionExpiredPage.scala.html
+++ b/app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/views/SessionExpiredPage.scala.html
@@ -14,17 +14,13 @@
  * limitations under the License.
  *@
 
-@import uk.gov.hmrc.nationalimportdutyadjustmentcentrefrontend.controllers.routes.LanguageSwitchController._
 @import uk.gov.hmrc.nationalimportdutyadjustmentcentrefrontend.config.AppConfig
-@import uk.gov.hmrc.hmrcfrontend.views.html.components.HmrcLanguageSelect
-@import uk.gov.hmrc.hmrcfrontend.views.viewmodels.language.LanguageSelect
 
-@this(hmrcLanguageSelect: HmrcLanguageSelect)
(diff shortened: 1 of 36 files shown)


app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/views/SessionExpiredPage.scala.html
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *@
+
+@this(layout: Layout)
+
+@()(implicit request: Request[_], messages: Messages)
+
+@layout(pageTitle = Some(messages("session_expired.title"))) {
+    <h1 class="govuk-heading-xl">@messages("session_expired.title")</h1>
```
crosslandwa, line 22: CLARIFICATION: I assume we're going to wait to break out the GOV Uk component twirl templates until we've got some more involved UI to implement?

i.e. implementing that as something like

```twirl
  @govuk.title(@messages("session_expired.title"))
```

SteveSugden: If there are components for basic "content" like this then I think we should use them.  I'll take a look.

crosslandwa: I can't see any mention of `govuk-heading` in [play-frontend-govuk](https://github.com/hmrc/play-frontend-govuk/search?q=%22govuk-heading%22) so i'm assuming there isn't  "component" for this low level content

I suppose we _could_ add one in this repo (though I'm happy that adding a layer of abstraction to hide adding a single class to a single html element is probably overkill)

crosslandwa (approved):

crosslandwa (approved):

PR #10, opened 2021-01-12
NF-78 - Compare new and original implementation

SteveSugden:


crosslandwa (approved):
Much prefer this approach over the "JSON blob" version

PR #12, opened 2021-01-13
NF-80 - CYA page

SteveSugden:


app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/navigation/Navigator.scala
```diff
@@ -0,0 +1,36 @@
...
+@Singleton
+class Navigator @Inject() () {
+
+  private val normalRoutes: Page => UserAnswers => Call = {
+    case ClaimTypePage => _ => controllers.makeclaim.routes.CheckYourAnswersController.onPageLoad()
+    case _             => _ => controllers.routes.StartController.onPageLoad()
+  }
+
+  def nextPage(page: Page, userAnswers: UserAnswers): Call =
+    normalRoutes(page)(userAnswers)
```
crosslandwa, line 34: CLARIFICATION: why do this with currying rather than a function that takes two parameters

i.e. something like
```scala
private val normalRoutes: (Page, UserAnswers) => Call = { ... }

...
def nextPage(page: Page, userAnswers: UserAnswers): Call = {
  normalRoutes(page, userAnswers)
}
```

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/views/Layout.scala.html
```diff
@@ -15,23 +15,32 @@
...
+        includeBackLink: Boolean = true
 )(contentBlock: Html)(implicit request: Request[_], messages: Messages)
 
 @beforeContentBlock = {
     @if(appConfig.welshLanguageSupportEnabled) {@languageSelect()}
+
+    @if(includeBackLink) {
+        @govukBackLink(BackLink(
+            classes = "govuk-!-display-none-print",
+            href = "javascript:history.back()",
```
crosslandwa, line 40: CLARIFICATION: Is implementing back with javascript OK from an accessibility (and/or users without javascript available) point of view?

I've seen this in other services so I _assume_ its been OK'd somewhere... The design system also says
> Where possible, ensure it works even when JavaScript is not available.

I'm happy for us to flag this up to address later (e.g. by having the server code generate the appropriate `href` attribute for the anchor element)

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/views/components/backLink.scala.html
```diff
@@ -0,0 +1,26 @@
...
+ *@
+
+@import uk.gov.hmrc.govukfrontend.views.html.components._
+
+@this(govukBackLink: GovukBackLink)
+
+@(messageKey: String = "site.back")(implicit messages: Messages)
+
+@govukBackLink(BackLink(
+    href = "javascript:history.back()",
```
crosslandwa, line 24: CLARIFICATION: Is implementing back with javascript OK from an accessibility (and/or users without javascript available) point of view?

I've seen this in other services so I assume its been OK'd somewhere... The design system also says

Where possible, ensure it works even when JavaScript is not available.

I'm happy for us to flag this up to address later (e.g. by having the server code generate the appropriate href attribute for the anchor element)

SteveSugden: This file is not needed.  I'll delete.

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/views/Layout.scala.html
```diff
@@ -15,23 +15,32 @@
...
-        scriptsBlock: Option[Html] = None
+        scriptsBlock: Option[Html] = None,
+        includeBackLink: Boolean = true
 )(contentBlock: Html)(implicit request: Request[_], messages: Messages)
 
 @beforeContentBlock = {
     @if(appConfig.welshLanguageSupportEnabled) {@languageSelect()}
+
+    @if(includeBackLink) {
+        @govukBackLink(BackLink(
```
crosslandwa, line 38: CLARIFICATION: Should this be `@components.backLink` to use our custom backlink component defined in `backLink.scala.html`?

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/views/components/backLink.scala.html
SteveSugden: To answer your comment.  I've seen services do both - use javascript and also try to implement the back-link logic in the server.    The later can become complex as you might imagine but can be done in a similar way to the Navigator.

The design system also says - "If this is not possible, you should hide the back link when JavaScript is not available" - so we could also do this.

I'd flag it for later discussion.    I just wanted to add a link as it appears on the prototype.

crosslandwa: > This file is not needed. I'll delete.

OK

> I'd flag it for later discussion. I just wanted to add a link as it appears on the prototype.

Yep, makes sense

> hide the back link when JavaScript is not available

I'd be in favour of doing this (unless someone tells us a good reason not to, e.g. during an assessment).

I'm not really comfortable with us writing business logic to generate "back URLs" that really send the user/browser _forward_ and cause the the browsers view of "back/history" to differ from the servers view.... (i.e. the 'back' button in the user's browser does something different to the "back" button in the page). In favour of side-stepping this if possible...

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/views/Layout.scala.html
crosslandwa: Resolving based on you other comment to delete `backLink.scala.html`

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/navigation/Navigator.scala
SteveSugden: That's how NDRC / scaffold project does it.  But I've implemented your suggestion as its a bit neater.

crosslandwa (approved):
Great!

PR #14, opened 2021-01-15
NF-81 - Tell user claim has been submitted successfully

SteveSugden:


app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/controllers/actions/DataRequiredAction.scala
```diff
@@ -21,18 +21,20 @@ import play.api.mvc.Results.Redirect
...
+class DataRequiredActionImpl @Inject() (val sessionRepository: UserAnswersRepository)(implicit
   val executionContext: ExecutionContext
 ) extends DataRequiredAction {
 
   override protected def refine[A](request: IdentifierRequest[A]): Future[Either[Result, DataRequest[A]]] =
     sessionRepository.get(request.identifier).map {
       case None =>
-        Left(Redirect(routes.SessionExpiredController.onPageLoad()))
+        Left(Redirect(routes.StartController.start()))
+      case Some(userAnswers) if userAnswers.claimReference.nonEmpty =>
```
crosslandwa, line 36: CLARIFICATION: My reading of this is that

- if the user tries to access a page that requires data
- and they have a ref number (i.e. they have submitted a claim)
- then redirect to start

Can you confirm my understanding that this means:
 -  when a user who has submitted a claim tries to access _any_ page with the `DataRequiredAction` configured they'll get sent to start (if they have completed a claim)
 - the confirmation page _cannot user_ the `DataRequiredAction`
 - users will not be able to start another claim until they log out or the mongo data expires

_I want to understand what constraints/limitations we'll have in place - i'm not advocating that we change the behaviour at this point_

SteveSugden: Your understanding is correct.  Except for the last point "users will not be able to start another claim until they log out or the mongo data expires".    Users _can_ start another application immediately.  The `DataRetrievalAction` used on the first page will ensure they get a new `UserAnswers` object created.

And the behaviour of `DateRequiredAction` is the reason it can't be used on the confirmation page.   Could create a new action that _only_ adds the claim ref to the request for use on the confirmation page but didn't think this was worth it.

crosslandwa (approved):

PR #16, opened 2021-01-18
BAU - read/write lastUpdated as date to ensure mongo ttl index works

SteveSugden:


app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/models/JsonFormats.scala
```diff
@@ -0,0 +1,42 @@
...
+
+package uk.gov.hmrc.nationalimportdutyadjustmentcentrefrontend.models
+
+import java.time.{Instant, LocalDateTime, ZoneOffset}
+
+import play.api.libs.json._
+
+object JsonFormats {
+
+  val formatLocalDateTime: OFormat[LocalDateTime] = new OFormat[LocalDateTime] {
```
crosslandwa, line 25: OBSERVATION: I think a name like `formatLocalDateTimeToEpoch` would make its usage within `UserAnswers` more obvious

crosslandwa (approved):

PR #17, opened 2021-01-18
BAU - sign out link in header

SteveSugden:


app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/controllers/SignOutController.scala
```diff
@@ -0,0 +1,49 @@
...
+  mcc: MessagesControllerComponents,
+  identify: IdentifierAction,
+  userAnswersRepository: UserAnswersRepository,
+  signedOutPage: SignedOutPage
+)(implicit ec: ExecutionContext)
+    extends FrontendController(mcc) with I18nSupport {
+
+  val signOut: Action[AnyContent] = identify.async { implicit request =>
+    userAnswersRepository.delete(request.identifier) map { _ =>
+      Redirect(controllers.routes.SignOutController.signedOut()).withNewSession
```
crosslandwa, line 41: OBSERVATION: The signOut Url is probably best extracted to config and updated to point to `/bas-gateway` as per

https://github.com/hmrc/national-duty-repayment-center-frontend/blob/89f7eb0a8d88f367d6b2506f9c9b83f830332102/conf/application.conf#L148


I think this also means we won't need to create a `SignedOutPage.scala.html` view or the `/signed-out ` application route

crosslandwa (changes requested):
Sign out URL needs updating to point to `/bas-gateway`

crosslandwa (approved):

PR #18, opened 2021-01-19
BAU - refactor caching of user data

SteveSugden:


app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/controllers/makeclaim/CheckYourAnswersController.scala
```diff
@@ -19,37 +19,48 @@ package uk.gov.hmrc.nationalimportdutyadjustmentcentrefrontend.controllers.makec
...
-    claimService.submitClaim(request.userAnswers) flatMap { response =>
-      val updatedCache = request.userAnswers.copy(claimReference = response.result)
-      userAnswersRepository.set(updatedCache) map {
-        _ => Redirect(routes.ConfirmationController.onPageLoad())
-      }
+    request.cacheData.answers match {
+      case Some(answers) =>
+        val claim = CreateClaimRequest(request.internalId, answers)
+        connector.submitClaim(claim) flatMap { response =>
+          repository.set(request.cacheData.copy(answers = None, createClaimResponse = Some(response))) map {
```
crosslandwa, line 55: Nice! I like that the answers are being cleared down explicitly here.

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/services/CacheDataService.scala
```diff
@@ -0,0 +1,40 @@
...
+  def getOrCreateAnswers(implicit request: IdentifierRequest[_]): Future[UserAnswers] =
+    repository.get(request.identifier) map { data =>
+      data.flatMap(_.answers).getOrElse(new UserAnswers())
+    }
+
+  def updateAnswers(update: UserAnswers => UserAnswers)(implicit request: IdentifierRequest[_]): Future[UserAnswers] =
+    repository.get(request.identifier) flatMap { maybeData =>
+      val data: CacheData             = maybeData.getOrElse(CacheData(request.identifier))
+      val updatedAnswers: UserAnswers = update(data.answers.getOrElse(UserAnswers()))
+      repository.set(data.copy(answers = Some(updatedAnswers))) map { _ => updatedAnswers }
```
crosslandwa, line 37: CLARIFICATION: I assume we want `lastUpdated` be updated (to _now_) in the `CacheData` instance we write to Mongo

Will `data.copy` create a new instance with `lastUpdated` appropriately set (to now) _if_ `data` represents the data we've just read from mongo?

crosslandwa (changes requested):
Theres one CLARIFICATION i'd like answering before I can approve

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/services/CacheDataService.scala
SteveSugden: Updating `lastUpdated` is still the responsibility of the `CacheDataRepository`.   The Service is just a convenience for an "Answers" controller to get and update UserAnswers without having to deal with the more complex CacheData model that we now have.

From CacheDataRepository - 
```
  def set(data: CacheData): Future[Option[CacheData]] =
    super.findAndUpdate(
      Json.obj("id" -> data.id),
      Json.toJson(data copy (lastUpdated = LocalDateTime.now)).as[JsObject],
      upsert = true
    ).map(_.value.map(_.as[CacheData]))
```

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/controllers/makeclaim/CheckYourAnswersController.scala
SteveSugden: Yes.  I pondered about doing this sort of thing in a method on the service, but decided that the CYA controller really should be responsible for maintaining / clearing the state in this instance.

crosslandwa (approved):

PR #19, opened 2021-01-21
NF-107 - file upload via Upscan.  Functionally complete without tests…

SteveSugden:
… for preview


app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/controllers/makeclaim/UploadFormController.scala
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  def showResult(uploadId: UploadId): Action[AnyContent] = identify.async { implicit request =>
+    uploadProgressTracker.getUploadResult(uploadId) flatMap {
+      case Some(successUpload: UploadedFile) =>
+        data.updateAnswers(answers => addUpload(answers, successUpload)) map {
+          updatedAnswers => Redirect(navigator.nextPage(UploadPage, updatedAnswers))
+        }
+      case Some(failed: Failed) =>
+        Future(Redirect(controllers.makeclaim.routes.UploadFormController.showError(failed.errorCode)))
```
crosslandwa, line 69: CLARIFICATION: Why is there a re-direct to the error page?

crosslandwa (dismissed):
LGTM - I'm interested in understanding why we're implementing the re-direct on error (because I'm curious, not because there's any reason to change it)

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/controllers/makeclaim/UploadFormController.scala
SteveSugden: Probably easier to demo.  The error page is used in two situations.   Firstly when the initial post to upscan fails  and upscan redirects the user to our /error endpoint.    The second is when the scanning fails.  We eventually detect this after the callback has happened and I'm treating the two types of "error" in the same way.

crosslandwa (approved):

PR #23, opened 2021-01-25
NF-146 - add JS fix to prevent 'confirm form re-submission' issue

SteveSugden:
JS file from paye-p11d-frontend project

### DIFF
--- a/app/assets/javascripts/nidacfrontend.js
+++ b/app/assets/javascripts/nidacfrontend.js
@@ -1,0 +1,24 @@
+window.GOVUKFrontend.initAll();
+window.HMRCFrontend.initAll();
+
+// =====================================================
+// Back link mimics browser back functionality
+// =====================================================
+// store referrer value to cater for IE - https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/10474810/  */
+const docReferrer = document.referrer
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/nidacfrontend.js
```diff
@@ -0,0 +1,24 @@
...
+// =====================================================
+// store referrer value to cater for IE - https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/10474810/  */
+const docReferrer = document.referrer
+
+// prevent resubmit warning
+if (window.history && window.history.replaceState && typeof window.history.replaceState === 'function') {
+  window.history.replaceState(null, null, window.location.href);
+}
+
+backLink = document.getElementById("back-link")
```
crosslandwa, line 15: OBSERVATION: This selector won't currently match anything

Our back links are being rendered as
```html
<a href="javascript:history.back()" class="govuk-back-link govuk-!-display-none-print">Back</a>
```

I'd suggest we add the ID (to be aligned with the "scaffold generated" JS code) rather than changing the selector (e.g. to `document.querySelector('a.govuk-back-link')`)

crosslandwa (approved):

PR #25, opened 2021-01-26
NF-159 Supply x-correlation-id

crosslandwa:
Derive value from incoming x-request-id (for consistency in tracing)


SteveSugden (approved):
Does the Upscan connector need this as well?  In which case it could be factored out into a Trait?

Also, I'd just do this in-line rather than create a private method e.g.
```
Seq("X-Correlation-Id" -> hc.requestId.map(_.value).getOrElse(UUID.randomUUID().toString))
```
Or if you keep it like this, make the hc an implicit arg.  Then you can get some scala magic where correlationId looks like a variable.
```
Seq("X-Correlation-Id" -> correlationId)
```

crosslandwa:
> Does the Upscan connector need this as well?  In which case it could be factored out into a Trait?

That will be (via the `HeaderCarrier`) setting an `x-request-id` which satisfies our tracing requirements

SteveSugden (approved):

PR #26, opened 2021-01-27
NF-101 - bank details

SteveSugden:


app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/forms/mappings/Validation.scala
```diff
@@ -0,0 +1,26 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package uk.gov.hmrc.nationalimportdutyadjustmentcentrefrontend.forms.mappings
+
+import scala.util.matching.Regex
+
+object Validation {
+
```
SteveSugden, line 22: The UI doesn't allow the entry of spaces or dashes  (numbers only) but I've used the same regex expression (and UI number input field) as NDRC

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/views/makeclaim/CheckYourAnswersPage.scala.html
```diff
@@ -38,7 +38,7 @@
         id = "claim_section",
         heading = Some(messages("check_answers.claim.title")),
         summaryListRows = Seq(
-            SummaryListRow(
+            Some(SummaryListRow(
```
crosslandwa, line 41: OBSERVATION: I find the use of `Some` here misleading - it suggests the `claimType` answer is optional - I understand its been added to support mapping the `Option[BankDetails]` to a `SummaryListRow`

I think i'd prefer the `summaryList.scala.html` to continue to take a `Seq[SummaryListRow]` and have the mapping of the `bankDetails` resolve to an empty `Seq`, something like

```
claim.bankDetails
  .map(bankDetails => Seq(SummaryListRow(...)))
  .getOrElse(Seq.empty)
```

(I'm happy for this suggestion to be ignored - I can see how it might become unwieldily when we have a mix of required/optional rows under a given summary list)

SteveSugden: It does indeed get messy where a section has a mixture of required and optional answers.  I had this problem in Exports and didn't find an easier way of building up a Seq(SummayListRow) fluidly on the view.

crosslandwa (approved):

PR #34, opened 2021-02-03
NF-175 - Repayment calculation - input fields

SteveSugden:


app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/forms/DutyPaidFormProvider.scala
```diff
@@ -0,0 +1,48 @@
...
+
+class DutyPaidFormProvider @Inject() extends Mappings {
+
+  def apply(): Form[DutyPaid] = Form(
+    mapping(
+      "actuallyPaid" -> text("dutyPaid.actual.error.required")
+        .verifying(
+          firstError(
+            regexp(Validation.dutyPattern, "dutyPaid.error.invalid"),
+            greaterThanZero("dutyPaid.actual.error.required")
```
crosslandwa, line 33: Given this uses `firstError`, will the regexp validation kick in if a leading `-` is supplied (preventing the `greaterThanZero` error from being surfaced)?

SteveSugden: It will.  `greatThanZero` is really filtering out `zero` and reporting that the same as a missing value (according to the requirements).   
Maybe that could be built into the regex by requiring a none-zero first character?  But what about "0.50"?   Actually, the regex is taken from the PEGA schema but modified to exclude leading `-`.    Could leave the PEGA regex un-touched and change the `dutyPaid.actual.error.required` message to something that indicates that a value greater than zero is required.

SteveSugden: Updated sightly.  The regex check now matches the PEGA schema (and will allow negative numbers), the "greaterThanZero" enforces the business requirements.

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/models/UserAnswers.scala
```diff
@@ -25,6 +25,7 @@ final case class UserAnswers(
   importerAddress: Option[Address] = None,
   claimType: Option[ClaimType] = None,
   reclaimDutyTypes: Option[Set[ReclaimDutyType]] = None,
+  reclaimDutyPayments: Map[String, DutyPaid] = Map.empty,
```
crosslandwa, line 28: CLARIFICATION: Are the keys in this `Map` of type `String` rather than `ReclaimDutyType` to make (de-)serialisation to/from Mongo more straightforward?

SteveSugden: Exactly

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/models/DutyPaid.scala
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package uk.gov.hmrc.nationalimportdutyadjustmentcentrefrontend.models
+
+import play.api.libs.json.{Json, OFormat}
+
+case class DutyPaid(actuallyPaid: String, shouldHavePaid: String) {
+  val dueAmount: BigDecimal = BigDecimal.apply(actuallyPaid) - BigDecimal.apply(shouldHavePaid)
```
crosslandwa, line 22: OBSERVATION: I think this should configure the `BigDecimal` instance to coerce to a 2dp representation to make the fact it handles pounds and pence explicit (at the moment that business rule is buried in the regex validation, which might be easy to miss/overlook)

test/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/models/eis/ClaimSpec.scala
```diff
@@ -0,0 +1,63 @@
...
+import uk.gov.hmrc.nationalimportdutyadjustmentcentrefrontend.models.ReclaimDutyType.{Customs, Other, Vat}
+import uk.gov.hmrc.nationalimportdutyadjustmentcentrefrontend.models.exceptions.MissingUserAnswersException
+
+class ClaimSpec extends UnitSpec with TestData {
+
+  "Claim" should {
+
+    "error" when {
+
+      "reclaim Customs Duty specified but no calculations present" in {
```
crosslandwa, line 30:
```suggestion
      "reclaim Customs Duty specified but no Customs duty payment details present" in {
```

crosslandwa (dismissed):

crosslandwa (approved):

PR #42, opened 2021-02-08
NF-194 - Auto sign-out when time expires

SteveSugden:


app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/controllers/KeepAliveController.scala
```diff
@@ -0,0 +1,42 @@
...
+@Singleton
+class KeepAliveController @Inject() (
+  mcc: MessagesControllerComponents,
+  identify: IdentifierAction,
+  cacheData: CacheDataService
+)(implicit ec: ExecutionContext)
+    extends FrontendController(mcc) with I18nSupport {
+
+  val keepAlive: Action[AnyContent] = identify.async { implicit request =>
+    cacheData.getAnswers map { _ =>
```
SteveSugden, line 37: Ensures that the users cached data does not expire

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/services/CacheDataService.scala
```diff
@@ -34,7 +34,7 @@ class CacheDataService @Inject() (repository: CacheDataRepository)(implicit ec:
     }
 
   def getAnswers(implicit request: IdentifierRequest[_]): Future[UserAnswers] =
-    getCacheData map (_.answers)
+    updateAnswers(answers => answers)
```
SteveSugden, line 37: Updates the 'lastUpdated' timestamp on the cached data.  Ensures that the users cache does not expire if they just keep hitting refresh to avoid getting logged out.

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/views/Head.scala.html
```diff
@@ -14,15 +14,28 @@
...
 
-@(headBlock: Option[Html] = None)
-@headBlock
+@this(hmrcTimeoutDialog: HmrcTimeoutDialog, appConfig: AppConfig)
 
-<script src='@routes.Assets.versioned("lib/govuk-frontend/govuk/all.js")'></script>
-<script src='@routes.Assets.versioned("lib/hmrc-frontend/hmrc/all.js")'></script>
-<script src='@controllers.routes.Assets.versioned("javascripts/nidacfrontend.js")'></script>
+@(headBlock: Option[Html] = None)(implicit request: Request[_], messages: Messages)
+@headBlock
```
SteveSugden, line 25: Moved to correct location according to play-frontend-hmrc guidance

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/views/Layout.scala.html
```diff
@@ -52,29 +52,35 @@
 }
 
 @signOutHref = @{
-    if(request.isInstanceOf[IdentifierRequest[_]] || request.isInstanceOf[DataRequest[_]])
```
SteveSugden, line 55: DataRequest no longer used - deleted

conf/application.conf
```diff
@@ -138,7 +138,7 @@ features.welsh-language-support = false
 
 mongodb {
   uri                 = "mongodb://localhost:27017/"${appName}
-  timeToLiveInSeconds = 900
+  timeToLiveInSeconds = 960
```
SteveSugden, line 141: Make this 60 secs longer than session timeout to avoid race-condition between user "saving" their session put the cache being deleted anyway.

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/services/CacheDataService.scala
crosslandwa: OBSERVATION:  This is a side-effect

I'd rather the `KeepAliveController` did

```
cacheData.getAnswers map { answers =>
  cacheData.update(answers) map {
    Ok("Ok")
  }
}
```

SteveSugden: Lol.  I did that originally until I realised that refreshing a page to "keep alive" was not updating the cache.

conf/application.conf
crosslandwa: Can that be described via:

```suggestion
  timeToLiveInSeconds = ${timeoutDialog. timeoutSeconds} + 60
```

SteveSugden: Can you do that?  I'll try it.

crosslandwa (changes requested):
LGTM

I'd like the side-effect from `CacheData::getAnswers` removed though (does something none obvious, and will increase load on Mongo as every GET page will do and read _and a write_ to the DB)

conf/application.conf
crosslandwa: > Can you do that? 

Ah, maybe not. I was thinking of referencing other bits of config (which you can done) and hadn't really considered if computations using those references were possible...

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/services/CacheDataService.scala
crosslandwa: Having spoken through this, i'm happy that updating `userAnswers.lastUpated` on read is a sensible way to ensure users sessions don't time out if they sit refreshing a given page

crosslandwa (approved):

crosslandwa (dismissed):

crosslandwa (approved):

PR #45, opened 2021-02-09
NF-139 - Show/view documents I uploaded

SteveSugden:


test/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/controllers/makeclaim/UploadFormSummaryControllerSpec.scala
```diff
@@ -0,0 +1,89 @@
...
+
+  }
+
+  "POST" should {
+
+    "redirect to next question" in {
+
+      val result = controller.onSubmit()(postRequest())
+      status(result) mustEqual SEE_OTHER
+      redirectLocation(result) mustBe Some(navigator.nextPage(UploadSummaryPage, emptyAnswers).url)
```
crosslandwa, line 85: OBSERVATION: I think i'd prefer this test to be explicit about what the next page should be (as per line 74) - this test could fail if the Navigator implementation/rules changed

crosslandwa (approved):
One OBSERVATION (which you can choose to ignore)

test/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/controllers/makeclaim/UploadFormSummaryControllerSpec.scala
SteveSugden: This test will be OK if/when navigator changes.   We are just asserting we go to the "next" page.  The actual page is up to the navigator.  I don't think any of the controller tests should be explicitly testing what the next page is.

The test at line 74 is different.   Getting to the "Upload" page is now directly controlled by the "UploadSummary" controller (based on the presence of existing uploaded files and in future, if the user selects that they want to upload another file)

PR #53, opened 2021-02-12
NF-180 - are you the importer or representative

SteveSugden:


crosslandwa (approved):
LGTM - as discussed we'll raise a separate ticket to change (if required) where in the `EisCaseRequest` we put the supplied `claim.contactDetails` - should this be conditional based on the `RepresentationType`?

PR #56, opened 2021-02-16
NF-215 - who do you want to repay?

SteveSugden:


app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/models/eis/EISCreateCaseRequest.scala
```diff
@@ -66,7 +67,7 @@ object EISCreateCaseRequest {
         EntryNumber = claim.entryDetails.entryNumber,
         EntryDate = eisDateFormatter.format(claim.entryDetails.entryDate),
         DutyDetails = claim.reclaimDutyPayments.map(entry => DutyDetail(entry._1, entry._2)).toSeq,
-        PayTo = "Importer", // TODO NF-204 - hard code values
+        PayTo = claim.repayTo.getOrElse(Importer).toString,
```
crosslandwa, line 70: OBSERVATION: I think i'd _prefer_ the defaulting to `Importer` to live in `Claim` (instead of it defaulting to `None` there) - it feels closer to "our business logic" than "logic to satisfy the PEGA schema"

Happy for this suggestion to be ignored (and I appreciate not having the `None` type in the `Claim` may have knock on effects in the CYA page)

crosslandwa (approved):

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/models/eis/EISCreateCaseRequest.scala
SteveSugden: It would have an impact on CYA.   And I think, on reflection, that it is a PEGA concern (it being mandated by the PEGA schema).  6 of 1 I guess.

PR #60, opened 2021-02-19
NF-227 - importer contact details

SteveSugden:


app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/models/Claim.scala
```diff
@@ -36,6 +36,7 @@ case class Claim(
   repayTo: Option[RepayTo],
   importerHasEoriNumber: Option[Boolean],
   importerEoriNumber: Option[EoriNumber],
+  importerContactDetails: Option[ImporterContactDetails],
```
crosslandwa, line 39: OBSERVATION: The modelling could be improved here IMO. If the `representationType 
 == representative` then `importerContactDetails` is not optional
 
 I suggest one of
 - creating `RepresentativeClaim` and `ImporterClaim` types
 - updating `Claim` to have a `Option[DetailsOfImporterBeingRepresented]` field, and within `DetailsOfImporterBeingRepresented` the `ImporterContactDetails` can be modelled as mandatory

Also, do we _need_ `Claim. importerHasEoriNumber` or can we infer it from the presence of `importerEoriNumber`? (I see how this change _may_ make the CYA page complicated....)

SteveSugden: Your main suggestion is a good one.   We now have a number of "optional not optional" fields for collecting the importer stuff for a representative claim.     Might be better as a separate PR?  I believe this is the last of the "importer" pages (for now).

Regarding `importerHasEoriNumber` - it's value _could_ be inferred.  But I've found that it is better to have the answer to each question persisted separately.  It gets complicated when the user changes the answer to this question.  As it stands if they change from yes to no we don't have to remove the EORI number (just ignore it).

SteveSugden: Sorry - just realised you are talking about Claim not UserAnswers.   Yes, we can probably do without importerHasEoriNumber.    (Not enough coffee)

crosslandwa: >  I've found that it is better to have the answer to each question persisted separately. It gets complicated when the user changes the answer to this question.

The answers to `importerHasEoriNumber` and `importerEoriNumber ` should be persisted in `UserAnswers` and I agree, keeping those answers explicit in there will help manage complexity.

I'm pondering if `importerHasEoriNumber` _needs_ to be pushed from `UserAnswers` to `Claim`, as `Claim.importerHasEoriNumber` is _only_ used to generate the "yes"/"no" on the CYA page for "does the importer have an eori" question...

(happy for this to be ignored, just a thought/observation)

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/models/eis/ImporterDetails.scala
```diff
@@ -17,26 +17,59 @@
...
+    case Representative =>
+      ImporterDetails(
+        claim.importerEoriNumber,
+        claim.importerContactDetails.getOrElse(throw new MissingUserAnswersException("Missing ImporterContactDetails"))
+      )
+    case Importer => ImporterDetails(claim.contactDetails, claim.claimantAddress)
+  }
+
+  // Applicant is the importer
+  def apply(contactDetails: ContactDetails, address: UkAddress): ImporterDetails = new ImporterDetails(
```
crosslandwa, line 45: OBSERVATION: I don't think these (public) apply functions are used anywhere other than within the `forClaim` method (I can only see test code using the case class directly) - making them private would 
 - clean up the API for the `ImporterDetails` object
 - allow them to be named so that the `// Applicant is the importer` comment is not necessary

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/models/Claim.scala
crosslandwa: re remodelling 

> Might be better as a separate PR?

I agree - can we spin up another ticket/PR straight off the back of this one to explore if the modelling can be improved

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/models/eis/ImporterDetails.scala
SteveSugden: Good idea.  The comments were for me so that I used the right one.   But a private method(s) would be better

crosslandwa (approved):

PR #62, opened 2021-02-19
BAU - refactor Claim to extract all importer information

SteveSugden:
All information that is only relevant for a representative making a claim is moved to `ImporterBeingRepresentedDetails`


app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/models/eis/EISCreateCaseRequest.scala
```diff
@@ -69,7 +69,7 @@ object EISCreateCaseRequest {
         EntryNumber = claim.entryDetails.entryNumber,
         EntryDate = eisDateFormatter.format(claim.entryDetails.entryDate),
         DutyDetails = claim.reclaimDutyPayments.map(entry => DutyDetail(entry._1, entry._2)).toSeq,
-        PayTo = claim.repayTo.getOrElse(Importer).toString,
+        PayTo = claim.importerBeingRepresentedDetails.map(_.repayTo).getOrElse(Importer).toString,
```
crosslandwa, line 72: OBSERVATION: On reflection @SteveSugden, this refactor highlights that defaulting to `Importer` is indeed a `EISCreateCaseRequest` concern rather than a `Claim` concern (as you surmised last time I raised this!)

crosslandwa (approved):

PR #64, opened 2021-02-22
NF-233 - basic BARS check

SteveSugden:


rorticus0502 (approved):

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/controllers/makeclaim/BankDetailsController.scala
```diff
@@ -60,12 +64,33 @@ class BankDetailsController @Inject() (
...
+  private def processBarsFailure(bankDetails: BankDetails, barsResult: BARSResult)(implicit
+    request: IdentifierRequest[_]
+  ) = {
+
+    val formWithErrors = barsResult match {
+      case bars if bars.accountNumberWithSortCodeIsValid != "yes" =>
+        form.fill(bankDetails).copy(errors =
+          Seq(FormError("accountNumber", "bankDetails.bars.validation.modCheckFailed"))
+        )
+      case _ => form.fill(bankDetails).copy(errors = Seq(FormError("", "bankDetails.bars.validation.failed")))
```
crosslandwa, line 86: CLAIRIFCATION: Under what circumstances does this `case` execute? 

(My lack of understanding of how `match/case` works mean I can't work out what would happen if `bars.accountNumberWithSortCodeIsValid == "yes"`)

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/models/Implicits.scala
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+package uk.gov.hmrc.nationalimportdutyadjustmentcentrefrontend.models
+
+object Implicits {
+
+  implicit class SanitizedString(unwrap: String) {
+    def stripSpacesAndDashes() = unwrap.replaceAll("""[ \-]""", "")
+
+    def leftPadToLength(length: Int, padChar: Char) = unwrap.reverse.padTo(length, padChar).reverse
```
crosslandwa, line 24: CLARIFICATION: Is the 2 x `reverse` a performance concern at all? (I know we made a change in the navigator to only reverse the (static) page order once, rather than every request)

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/controllers/makeclaim/BankDetailsController.scala
SteveSugden: This method shouldn't be called if there are no validation problems (currently that means accountNumberWithSortCodeIsValid = "yes") but if it _were_ called you would get the other "case" which is intended as a "there is a problem" catch-all

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/models/Implicits.scala
SteveSugden: I nicked those two implicit methods from BARS FE.  There's probably a more efficient way of coding this.  I'll have a play.

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/models/BankDetails.scala
```diff
@@ -17,9 +17,14 @@
...
 import play.api.libs.json.{Json, OFormat}
+import uk.gov.hmrc.nationalimportdutyadjustmentcentrefrontend.models.Implicits.SanitizedString
 
 case class BankDetails(accountName: String, sortCode: String, accountNumber: String)
 
 object BankDetails {
   implicit val format: OFormat[BankDetails] = Json.format[BankDetails]
+
+  def apply(accountName: String, sortCode: String, accountNumber: String): BankDetails =
+    new BankDetails(accountName, sortCode, accountNumber.leftPadToLength(8, '0'))
```
crosslandwa, line 28: CLARIFICATION: Does the left padding here ensure that:
- the value sent to BARS (for validation) is always 8 chars?
- the value sent to EIS/PEGA is always 8 chars?

crosslandwa (dismissed):

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/controllers/makeclaim/BankDetailsController.scala
crosslandwa: Ah, right, yes, we'd never get into `processBarsFailure` if `accountNumberWithSortCodeIsValid = "yes"`

Makes sense

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/models/BankDetails.scala
SteveSugden: Both.  BARS requires 8 digits and it seems reasonable to send to PEGA what we have had verified.

crosslandwa: 👍 

PEGA requires 8 digits too - just checking this change makes us compliant with that

crosslandwa (approved):

PR #71, opened 2021-02-25
BAU - create generic Navigator

SteveSugden:


app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/navigation/Navigator.scala
```diff
@@ -16,72 +16,25 @@
...
-
-  protected val enterImporterEori: CreateAnswers => Boolean = (answers: CreateAnswers) =>
-    isRepresentative(answers) && answers.doesImporterHaveEori
-
-}
-
-protected trait Ordering {
-
-  protected val nextPageFor: (Seq[P], Page, CreateAnswers) => Option[Page] = (pages, currentPage, userAnswers) =>
+  protected val nextPageFor: (Seq[P], Page, T) => Option[Page] = (pages, currentPage, userAnswers) =>
```
crosslandwa, line 37: `nextPageFor` can be private

```suggestion
  private val nextPageFor: (Seq[P], Page, T) => Option[Page] = (pages, currentPage, userAnswers) =>
```

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/navigation/Navigator.scala
```diff
@@ -99,4 +52,5 @@ protected trait Ordering {
     case s              => s.tail
   }
 
+  protected val always: Answers => Boolean = (_: Answers) => true
```
crosslandwa, line 55: OBSERVATION: Moving `always` out from the `Conditions` trait makes the intended usage of this method by extending classes ambiguous IMO

I'd prefer one of
- moving `always` into `CreateAnswerConditions` (and duplicating it in `AmendAnswerConditions` when that exists)
- creating some sort of `SharedConditions` trait that is available for use by both `CreateNavigator` and `AmendNavigator`

crosslandwa (approved):
LGTM

FWIW i'd prefer to do this without extension (for fear of the `Navigator` trait attracting more complexity in the future to cater for specific extending classes), but I appreciate that (for now) having `CreateNavigator` and `AmendNavigator` _only_ have to define a page order and the access conditions is pretty neat/easy to understand

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/navigation/Navigator.scala
SteveSugden: Is that an either/or suggestion?

crosslandwa: Yes, its an either/or (though the first is the simpler one to implement while we only have one class extending the `Navigator` trait)

crosslandwa (dismissed):

crosslandwa (approved):

PR #72, opened 2021-02-25
BAU - move create case form providers into new package

SteveSugden:


test/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/forms/AddressFormProviderSpec.scala
```diff
@@ -19,6 +19,7 @@ package uk.gov.hmrc.nationalimportdutyadjustmentcentrefrontend.forms
 import org.scalacheck.Gen
 import play.api.data.FormError
 import uk.gov.hmrc.nationalimportdutyadjustmentcentrefrontend.forms.behaviours.StringFieldBehaviours
+import uk.gov.hmrc.nationalimportdutyadjustmentcentrefrontend.forms.create.AddressFormProvider
```
crosslandwa, line 22: OBSERVATION: We should be moving `AddressFormProviderSpec.scala` (and the other form specs) into a `create` subpackage to match the form under test...

crosslandwa (changes requested):
Please move the FormSpecs into a `create` sub-package (for consistency with the forms they test)

test/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/forms/AddressFormProviderSpec.scala
SteveSugden: For the money, I would expect IDEA to have done that for me.
Done.

crosslandwa (approved):

test/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/forms/AddressFormProviderSpec.scala
crosslandwa: 💰 💰 💰

PR #73, opened 2021-02-25
NF-241 - amend journey case reference number

SteveSugden:


app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/pages/QuestionPage.scala
```diff
@@ -51,3 +38,5 @@ case object ClaimReasonPage            extends QuestionPage[ClaimReason]
 case object UploadSummaryPage          extends QuestionPage[Boolean]
 case object EntryDetailsPage           extends QuestionPage[EntryDetails]
 case object ItemNumbersPage            extends QuestionPage[ItemNumbers]
+
+case object CaseReferencePage extends QuestionPage[CaseReference]
```
crosslandwa, line 42: OBSERVATION: A comment here to distinguish the amend case pages from the create case pages would be useful

crosslandwa (approved):

PR #78, opened 2021-02-26
BAU - use AmendClaim model to validate required answers

SteveSugden:
- add default /amend endpoint to redirect to first amend claim question

### DIFF
--- a/conf/amendClaim.routes
+++ b/conf/amendClaim.routes
@@ -1,5 +1,6 @@
 
 # amend claim
+GET         /                                   uk.gov.hmrc.nationalimportdutyadjustmentcentrefrontend.controllers.amendclaim.AmendClaimController.start()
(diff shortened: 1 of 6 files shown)


conf/amendClaim.routes
```diff
@@ -1,5 +1,6 @@
 
 # amend claim
+GET         /                                   uk.gov.hmrc.nationalimportdutyadjustmentcentrefrontend.controllers.amendclaim.AmendClaimController.start()
```
rorticus0502, line 3: app.routes also has a / endpoint that points to the StartController

Can we have two?

rorticus0502: Oh I think I get it.
Ignore me.

rorticus0502 (approved):

PR #79, opened 2021-02-26
NF-259 - upload document for amend case journey

SteveSugden:


SteveSugden:
image
image

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/views/amendclaim/UploadProgressView.scala.html
```diff
@@ -0,0 +1,38 @@
...
+@import uk.gov.hmrc.nationalimportdutyadjustmentcentrefrontend.views.html.components.uploadBodyAmend
+@import uk.gov.hmrc.nationalimportdutyadjustmentcentrefrontend.views.html.components.uploadProgress
+@import uk.gov.hmrc.nationalimportdutyadjustmentcentrefrontend.viewmodels.NavigatorBack
+
+@this(layout: Layout, continueButton: components.continueButton,uploadBody: uploadBodyAmend, uploadProgress: uploadProgress)
+
+@(backLink: NavigatorBack)(implicit request: Request[_], messages: Messages)
+
+@refreshHeader = {
+    <meta http-equiv="refresh" content="1"/>
```
crosslandwa, line 29: OBSERVATION: I wonder how users of screen-readers will find the page automatically reloading every 1sec...

I wonder if we should be updating the page title to detail that an upload is in progress (as this _should_ be the first thing the screen reader reads)

crosslandwa: Same applies to the create-claim uploads as well, so i suggest we address this (if necessary) in a separate change

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/controllers/amendclaim/UploadFormController.scala
```diff
@@ -0,0 +1,118 @@
...
+  val navigator: AmendNavigator,
+  uploadFormView: UploadFormView,
+  uploadProgressView: UploadProgressView
+)(implicit ec: ExecutionContext)
+    extends FrontendController(mcc) with I18nSupport with Navigation[AmendAnswers] with FileUploading {
+
+  override val page: Page = UploadPage
+
+  protected val errorRedirectUrl =
+    appConfig.upscan.redirectBase + "/national-import-duty-adjustment-centre/amend/upload-additional-documents/error"
```
crosslandwa, line 59: CLARIFICATION: Is there a reason this can't be some incantation of

```
protected val errorRedirectUrl = errorCode: String => appConfig.upscan.redirectBase + routes.UploadFormController.onError(errorCode).url
```

I'm a bit wooly on where the `errorCode` comes from when Upscan redirects the user to the error page...

crosslandwa: Same applies for `makeclaim/UploadFormController`

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/views/amendclaim/UploadProgressView.scala.html
SteveSugden: Good question and good idea.

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/controllers/amendclaim/UploadFormController.scala
SteveSugden: The error code is appended to the errorRedirectUrl we give to upscan in the form of a query param.   I'll have a go at programmatically creating the url with an empty or dummy error code and then stripping off the query parameter

SteveSugden: I.e. we need to give `/national-import-duty-adjustment-centre/create/upload-supporting-documents/error` to upscan.  They append the following (as an example) `?key=e99cb31f-4b03-47bb-9746-d8ac2be4526a&errorCode=InvalidArgument&errorMessage=%27file%27+field+not+found&errorResource=NoFileReference&errorRequestId=SomeRequestId`

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/controllers/amendclaim/UploadFormController.scala
```diff
@@ -0,0 +1,118 @@
...
+        uploadInitialView(upscanInitiateResponse, answersWithJourneyID._1, None)
+      }
+    }
+  }
+
+  def onProgress(uploadId: UploadId): Action[AnyContent] = identify.async { implicit request =>
+    data.getAmendAnswersWithJourneyId flatMap { answersWithJourneyID =>
+      uploadProgressTracker.getUploadResult(uploadId, answersWithJourneyID._2) flatMap {
+        case Some(successUpload: UploadedFile) =>
+          processSuccessfulUpload(successUpload)
```
crosslandwa, line 76: OBSERVATION: After the `processSuccessfulUpload` completes should we be clearing the upload attempt from the `uploadProgressTracker`?

SteveSugden: We could do.  I don't think we need that information any more.

SteveSugden: But we don't "clean-up" unsuccessful upload attempts.

crosslandwa: Good point - I think letting mongo clean itself up based on data retention period is fine (given the expected volumes)

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/views/amendclaim/UploadProgressView.scala.html
crosslandwa: Raised a new ticket to investigate accessibility issues with the file upload pattern

crosslandwa (dismissed):

crosslandwa (approved):

PR #92, opened 2021-03-04
NF-285 - implement BARs checks with /metadata and /assess endpoints

SteveSugden:


app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/models/bars/AssessBusinessBankDetailsResponse.scala
```diff
@@ -0,0 +1,50 @@
...
+
+case class AssessBusinessBankDetailsResponse(
+  accountNumberWithSortCodeIsValid: String,
+  nonStandardAccountDetailsRequiredForBacs: String,
+  accountExists: String,
+  companyNameMatches: String
+) {
+  val validAccountAndSortCode: Boolean = accountNumberWithSortCodeIsValid == yes
+  val rollNotRequired: Boolean         = nonStandardAccountDetailsRequiredForBacs == no
+  val accountValid: Boolean            = Set(yes, indeterminate).contains(accountExists)
```
crosslandwa, line 34: OBSERVATION: This is interpretation of the `"yes"/"no"/"indeterminate"` values is "NIDAC business logic" - I think it should be pulled up and out of this `BarsResponse` object and into the `BarsResult`

crosslandwa: Or, put another way, I'd expect the `BarsResponse` case class to do nothing above and beyond deserialising the BARS API response (any interpretation of that response should be elsewhere)

test/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/base/BarsTestData.scala
```diff
@@ -0,0 +1,43 @@
...
+  BARSResult,
+  MetadataResponse
+}
+
+trait BarsTestData extends TestData {
+
+  // Metadata
+  val validMetaDataResponse        = MetadataResponse("M")
+  val noBacsMetaDataResponse       = MetadataResponse("N")
+  val noBacsCreditMetaDataResponse = MetadataResponse("M", Seq("CR"))
```
njmoorewitchcat, line 30: AMBER: We don't explicitly test the bacsOfficeStatus of A combintation (with and without CR).  Do we need to?

njmoorewitchcat (approved):

test/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/base/BarsTestData.scala
njmoorewitchcat, line 30: AMBER: We don't explicitly test the bacsOfficeStatus of A combintation (with and without CR). Do we need to?

SteveSugden: Given the definition of `val acceptsBacsPayments = bacsOfficeStatus != "N" && !disallowedTransactions.contains("CR")` I don't think so.  But I can add it if you want.

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/models/bars/AssessBusinessBankDetailsResponse.scala
SteveSugden: Is that a "style" thing?  I think keeping the logic about each call in each response is reasonable.  In fact, if I hadn't defined the responses as case classes I could have hidden the actual response fields - it's the boolean interpretations that matter.

test/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/base/BarsTestData.scala
njmoorewitchcat: I was thinking future developers might refactor that bacsOfficeStatus == "M", and the tests would still pass.

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/models/bars/AssessBusinessBankDetailsResponse.scala
crosslandwa: Yeah, just a style thing, totally happy for you to ignore this suggestion

njmoorewitchcat (approved):
LGTM

PR #93, opened 2021-03-04
NF-291 list document types and test

rorticus0502:


njmoorewitchcat (approved):
LGTM

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/views/components/uploadDocumentTypes.scala.html
```diff
@@ -28,6 +28,7 @@
   case Some(AntiDumping)   => List(E2, C88, Invoice)
   case Some(Preference)    => List(E2, C88, Invoice, PreferenceCertificate)
   case Some(Quota)         => List(E2, C88, Invoice, PreferenceCertificate)
+  case Some(AccountSales)  => List(E2, C88, Invoice, CertificateOfOrigin)
   case _                   => DocumentTypes.values.toList
```
SteveSugden, line 32: It might have been better if I had not been so defensive and included the catch-all `case _` - then it would have blown up when I added the new claim type and this change would not have been missed.

PR #99, opened 2021-03-05
Nf 292 postcode validation re revisited

rorticus0502:


SteveSugden (changes requested):
I'm not sure that this is the best approach.   Sanitising the input _before_ the form binding seems wrong.  Doubly wrong as you've had to do it twice.
I think it should be done in the form mapping (and would be even better if we had a single "postcode" mapping that the 2 form providers could share)

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/forms/mappings/Constraints.scala
```diff
@@ -33,6 +34,24 @@ trait Constraints {
...
+    Constraint {
+      case str if str.stripExternalAndReduceInternalSpaces().length > 8 =>
+        Invalid(errorKey)
+      case str if str.stripExternalAndReduceInternalSpaces().length < 5 =>
+        Invalid(errorKey)
+      case _ =>
+        Valid
+    }
+
+  protected def postcodeFormat(errorKey: String): Constraint[String] =
```
SteveSugden, line 47: You could use that new `regexp` method below and pass in the `stripExternalAndReduceInternalSpaces` method.

But having an explicitly name method for postcodes is good too.

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/models/eis/ImporterAddress.scala
```diff
@@ -30,4 +32,8 @@ case class ImporterAddress(
 
 object ImporterAddress {
   implicit val format: OFormat[ImporterAddress] = Json.format[ImporterAddress]
+
+  def apply(name: String, addressLine1: String, addressLine2: Option[String], city: String, postCode: String): Address =
```
SteveSugden, line 36: Don't need to apply the formatting here.  The `eis` models are constructed from "safe data" (i.e. our user-answer models)

test/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/forms/create/AddressFormProviderSpec.scala
```diff
@@ -106,48 +106,42 @@ class AddressFormProviderSpec extends StringFieldBehaviours {
...
-      form,
-      fieldName,
-      minLength = minLength,
-      lengthError = FormError(fieldName, lengthKey, Seq(minLength))
-    )
-
     "not bind strings with invalid characters" in {
       val result        = form.bind(Map(fieldName -> "P!24KF")).apply(fieldName)
-      val expectedError = FormError(fieldName, invalidKey, Seq(Validation.postcodePattern))
+      val expectedError = FormError(fieldName, invalidKey)
```
SteveSugden, line 124: Do you know why the regex is not part of the FormError now?   It's not important as we don't really use it, just curious.

SteveSugden (commented):
Much happier with this approach, thanks. 
Only thing I would recommend changing is the formatting on the "EIS" address model

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/forms/mappings/Constraints.scala
```diff
@@ -33,6 +34,24 @@ trait Constraints {
           .getOrElse(Valid)
     }
 
+  protected def postcodeLength(errorKey: String): Constraint[String] =
+    Constraint {
```
SteveSugden, line 38: We could have used the existing maxLength and minLength constriants by passing in a "transformation" function in a similar way to the regexp method.

But again, having a postcode specific method is ok.

SteveSugden (approved):

PR #100, opened 2021-03-05
Nf 290 cya tidy

njmoorewitchcat:


app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/views/components/subHeading.scala.html
```diff
@@ -16,4 +16,4 @@
 
 @(content: String)
 
-<h2 class="govuk-heading-m">@{Text(content).asHtml}</h2>
```
SteveSugden, line 19: Is this subHeading component used elsewhere (i.e. not on CYA) and if so us that spacing appropriate there as well?

conf/messages
```diff
@@ -254,7 +254,10 @@ check_answers.payment.title = Payment information
 check_answers.payment.payTo = Who is to be repaid?
 check_answers.payment.payTo.Importer = Importer
 check_answers.payment.payTo.Representative_of_importer = Representative
-check_answers.payment.bankDetails = Bank details
+check_answers.payment.bankDetails = Name on the account<br>Sort code<br>Account number
```
SteveSugden, line 257: I questioned this label when I originally did bank details.  IMO having three "labels" like this is not right.  Also having mark-up in the message is not ideal.   I personally would push back on the design but if we have to implement it like this then have three messages and do the formatting in the view

SteveSugden (approved):
LGTM - just a couple of clarifications/suggestions.

Have you checked the ATs are still ok?  They rely on the pseudo class names for the CYA sections (if you've not changed the names then all should be OK)

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/views/components/subHeading.scala.html
njmoorewitchcat: SubHeading is used in CreateCheckYourAnswers and AmendCheckYourAnswers only for the headers between sections.  If we decide to add it to other pages, we probably want to create a special CheckYourAnswers subHeading or something else then.  At the moment its safe.

conf/messages
njmoorewitchcat: I will get a steer from design team.  I agree 100% with you on the better 'Bank details'.  And yeah, brain dead when I implemented, your are correct, should be three seperate messages - If I need to change it based on the feedback, I will do that.

njmoorewitchcat:
> LGTM - just a couple of clarifications/suggestions.
> 
> Have you checked the ATs are still ok? They rely on the pseudo class names for the CYA sections (if you've not changed the names then all should be OK)

Will check them now - I have removed a section, so they may not be ok..

conf/messages
njmoorewitchcat: Changing this back - we will have a longer discussion with the design team on deviations from the prototype

SteveSugden (approved):

PR #101, opened 2021-03-06
NF-294 - convert from simple-reactivemongo to hmrc-mongo-play-27

SteveSugden:


app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/repositories/CacheDataRepository.scala
```diff
@@ -17,54 +17,46 @@
...
+      mongoComponent = mongoComponent,
       domainFormat = CacheData.formats,
-      idFormat = ReactiveMongoFormats.objectIdFormats
+      indexes = Seq(
+        IndexModel(
+          ascending("lastUpdated"),
+          IndexOptions().name("userAnswersExpiry").expireAfter(config.mongoTimeToLiveInSeconds, TimeUnit.SECONDS)
+        )
+      ),
+      replaceIndexes = true
```
SteveSugden, line 45: Might want to discuss this.  Can be expensive at start-up but I don't think our dbs with 15min ttl will ever contain many records.

SteveSugden: Also, we need to consider adding an explicit index for "id".  Mongo creates an index for the internal "_id_" it creates but this is not the user Id we do searches on.

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/repositories/UploadRepository.scala
```diff
@@ -79,42 +78,35 @@ object UploadDetails {
...
+      mongoComponent = mongoComponent,
       domainFormat = UploadDetails.format,
-      idFormat = ReactiveMongoFormats.objectIdFormats
+      indexes = Seq(
+        IndexModel(
+          ascending("created"),
+          IndexOptions().name("uploadExpiry").expireAfter(config.mongoTimeToLiveInSeconds, TimeUnit.SECONDS)
+        )
+      ),
+      replaceIndexes = true
```
SteveSugden, line 95: Same comment re" indexes.   We might need to add composite indexes for the two pairs of fields we search on.

njmoorewitchcat (approved):

crosslandwa (approved):

PR #103, opened 2021-03-09
Nf 254 gtm

rorticus0502:


rorticus0502:
> Seems reasonable. Is that pattern (head and body) copied from somewhere (another service or confluence instructions)? It's been a while since I did this sort of thing.

I looked at how route one embedded the gtm tags and followed similar approach

PR #113, opened 2021-03-10
NF-284 What Do You Want To Do Page added

njmoorewitchcat:


rorticus0502 (dismissed):

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/controllers/WhatDoYouWantToDoController.scala
```diff
@@ -0,0 +1,62 @@
...
+    Ok(whatDoYouWantToDoPage(form))
+  }
+
+  def onSubmit(): Action[AnyContent] = identify { implicit request =>
+    form.bindFromRequest().fold(
+      formWithErrors => BadRequest(whatDoYouWantToDoPage(formWithErrors)),
+      {
+        case ToDoType.NewClaim   => Redirect(createNavigator.nextPage(FirstPage, CreateAnswers()))
+        case ToDoType.AmendClaim => Redirect(amendNavigator.nextPage(FirstPage, AmendAnswers()))
+        case _                   => BadRequest(whatDoYouWantToDoPage(form))
```
SteveSugden, line 57: Don't think that line 57 could ever happen.  And if it did there are no errors on the form so the user wouldn't know what to do.

SteveSugden (approved):
LGTM

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/controllers/WhatDoYouWantToDoController.scala
njmoorewitchcat: Belt and Braces - I didn't want to leave a non default case out of the match

SteveSugden: Better to fail with an "unmatched case exception" than to partially handle the condition. But like I said, it can't happen ((unless someone added a new type)

PR #119, opened 2021-03-11
NF-298 - update service name

SteveSugden:


conf/messages
```diff
@@ -1,11 +1,11 @@
-service.name = National Import Duty Adjustment Centre
+service.name = Apply for return of import duty secured by deposit or guarantee
 service.homePageUrl = /national-import-duty-adjustment-centre
```
njmoorewitchcat, line 2: AMBER: Is this still the right URL for the homePageUrl?

rorticus0502 (dismissed):

conf/messages
SteveSugden: Not its not.  Should be RED!.    Not sure why that is defined in messages.  Will take a look

SteveSugden: Message key was not used.   Added when the project was generated I presume.  All urls used within the service are programatically derived.  But good spot & thank you.

njmoorewitchcat (approved):

PR #122, opened 2021-03-12
NF-298 - change refund to return in service name url

SteveSugden:


docs/NIDAC_Create_Claim.js
```diff
@@ -1,11 +1,11 @@
 // ==UserScript==
 // @name         NIDAC Create Claim AutoComplete
 // @namespace    http://tampermonkey.net/
-// @version      0.15
+// @version      0.16
 
 // @description  NIDAC Create Claim AutoComplete
 // @author       NIDAC Team
-// @match        http*://*/apply-for-refund-import-duty-paid-on-deposit-or-guarantee/*
+// @match        http*://*/apply-for-return-import-duty-paid-on-deposit-or-guarantee/*
```
njmoorewitchcat, line 8: RED: In the same directory docs/NIDAC_Auth_Stub.js has not been updated.

njmoorewitchcat (changes requested):
README.md has a refund reference that needs updating. 

Also does docs/NIDAC_Create_Claim.js need deleting > it has a reference as well (and possibly the mentioned AuthStuf)

njmoorewitchcat (approved):

PR #125, opened 2021-03-15
NF-318 - return amount summary screen change links

SteveSugden:


app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/navigation/CreateNavigator.scala
```diff
@@ -184,6 +190,6 @@ protected trait CreateHasAnsweredConditions {
   protected val bankDetailsAnswered: CreateAnswers => Boolean = _.bankDetails.nonEmpty
 
   protected val returnSummaryAnswered: CreateAnswers => Boolean = (answers: CreateAnswers) =>
-    answers.changePage != Some(CreatePageNames.dutyTypes)
+    answers.changePage.exists(page => !page.startsWith("duty"))
```
rorticus0502, line 193: the predicate here, using string startsWith feels fragile
It will work with the current set of pageNames

rorticus0502 (dismissed):

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/navigation/CreateNavigator.scala
SteveSugden: I know what you mean.  I'll create a set of "duty" pages instead

SteveSugden: Implemented less fragile check for "duty" pages

rorticus0502 (approved):

PR #127, opened 2021-03-15
BAU - goto missing question rather than start using navigator

SteveSugden:


crosslandwa:
This looks sensible to me

Out of interest, why aren't the duty/vat/other pages defined in the page names already?

SteveSugden:
Because until today, you couldn't "change" those values (via a change link)

PR #128, opened 2021-03-16
NF-331 - update ids of submit buttons

SteveSugden:


crosslandwa (dismissed):
OBSERVATION: Personally I'd prefer we spin up a `@submitButton` component that internally hard-codes the id attribute to `nidac-submit`

Similarly the existing `@continueButton` would internally hard-code the id (rather than accepting it as a parameter with a default value)

This preference is largely a matter of taste - i'd prefer "more components whose intended usage was more explicit" over "fewer components with more parameterisation options"

SteveSugden:
submitButton created

rorticus0502 (approved):

PR #131, opened 2021-03-17
NF-337 - page title on error condition

SteveSugden:


rorticus0502 (approved):

njmoorewitchcat (commented):
AMBER: Our PageView tests run an error scenario.  We should add an extra verification test there to confirm that the page title has "Error:" in front for all these pages.  This step may have caught the original issue.

    "display error when no choice is made" in {
      val errorView = view(form.bind(Map("claim-type" -> "")))
      errorView.getElementsByClass("govuk-error-summary__body").text() mustBe messages("claim_type.error.required")

      errorView().title() must startWith(messages("Error:"))

    }

SteveSugden:
Yes that would be a good addition to the "ViewMatchers".   Something that verifies that a page has an error summary with the expected text and also that the page title starts with Error.   Leave it with me

PR #138, opened 2021-03-23
NF-325 - support for EORI allow list

SteveSugden:


conf/application.conf
```diff
@@ -173,3 +173,14 @@ upscan {
 
 timeoutDialog.countdownSeconds = 120
 
+eori {
+    enrolments = [
+      "HMRC-CTS-ORG"
+    ],
+    allowList {
+      enabled = false
```
crosslandwa, line 181: CLARIFICATION: The acceptance tests require the allow list to be enabled (they include a `Scenario: user with non-allowed EORI cannot login`)

Does this config conflict with the needs of the acceptance tests?

SteveSugden: Yes.  It does.  Thanks.   I changed this local config after testing the AT.
Shall I enable the allow list in application.conf?  Or remove the AT?

njmoorewitchcat (approved):

conf/application.conf
crosslandwa: I was thinking of commenting against `IdentifierAction.scala` to request extracting/unit testing of the access logic

I think _removing the AT_ and replacing it with unit tests will reduce the overhead of managing this feature (and remove coupling between the AC suite and environment specific config)...

SteveSugden: Done - `AuthenticatedIdentifierActionSpec` enhanced with additional tests

crosslandwa (approved):

PR #143, opened 2021-03-24
Nf 334 send create claim audit

rorticus0502:


app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/models/create/CreateClaimAudit.scala
```diff
@@ -0,0 +1,46 @@
...
+  claimReason: ClaimReason,
+  reclaimDutyPayments: Map[String, DutyPaid] = Map.empty,
+  bankDetails: BankDetails,
+  importerContactDetails: Option[ImporterContactDetails] = None,
+  repayTo: Option[RepayTo] = None,
+  entryDetails: EntryDetails,
+  itemNumbers: ItemNumbers,
+  uploads: Seq[UploadedFile] = Seq.empty,
+  fileTransferResults: Seq[FileTransferResult] = Seq.empty,
+  importerEori: Option[EoriNumber] = None
```
SteveSugden, line 39: Is there a need for defaulting these optional and collection field values?  I don't think you are constructing this class anywhere and _not_ providing all of the values

rorticus0502: Good spot, thank you

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/services/ClaimService.scala
```diff
@@ -63,4 +74,36 @@ class ClaimService @Inject() (connector: NIDACConnector) {
 
   }
 
+  def audit(success: Boolean, claim: Claim, claimResponse: CreateClaimResponse)(implicit
+    hc: HeaderCarrier,
+    ec: ExecutionContext
+  ): Unit = {
+
+    val audit = new CreateClaimAudit(
```
SteveSugden, line 82: Observation:  Do you need the "new" keyword for a case class?   Also, you could create an "apply" method on a companion `CreateClaimAudit` object that does this construction.   So the code here could then be reduced to 
``` val audit = CreateClaimAudit(success, claim, claimResponse)```

File: test/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/base/TestData.scala [hunk 3]
```diff
@@ -147,4 +151,45 @@ trait TestData {
       uploadAnotherFile = Some(uploadAnotherFileAnswer)
     )
 
+  val claim: Claim = Claim(completeAnswers)
```
SteveSugden, line 154: You need to update from master - Claim now takes the user's EORI number

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/services/ClaimService.scala
rorticus0502: will do

test/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/base/TestData.scala
rorticus0502: will do

SteveSugden (approved):
Looks good

PR #145, opened 2021-03-25
BAU - proposed fix for changing bank details

SteveSugden:


crosslandwa:
OBSERVATION: I think this change makes sense (modelling importer/agent bank details separately), but I believe the proposed implementation (re-)introduces the behaviour where
- user selects `representationType = agent`
- user selects `repayTo = agent`
- user enters (agents) bank details
- user changes `repayTo = importer`
- user enters (importers) bank details
- user changes `repayTo = agent`
- user submits claim with "agent bank details", rather than "the most recently entered" bank details

We should confirm what the user expectation is in this scenario

rorticus0502:
Observation

I think we have some logic in the claim that decides what to do, based on RepresentationType and RepyTo
Now we have some in the Answers model.

Can we find a way to have it in one place?

SteveSugden:
It does behave as you've described.   But it is submitting the correct bank details rather the one most recently entered.  The bank details page distinguishes between "Your bank details" and "The importer's bank details" so I don't think that is a problem.

But I'm not a typical user, of course.

SteveSugden:
> Observation
> 
> I think we have some logic in the claim that decides what to do, based on RepresentationType and RepyTo
> Now we have some in the Answers model.
> 
> Can we find a way to have it in one place?

Possibly/probably.   Will investigate if we decide to go with this approach.

crosslandwa (approved):

PR #146, opened 2021-03-25
NF-338 - Missing answers on CYA

SteveSugden:
This is based on the prototype I started a couple of weeks ago.   Although that approach has been rejected (automatically directing user to first missing answer) the changes to the navigator are still required.


app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/controllers/amendclaim/CheckYourAnswersController.scala
```diff
@@ -45,11 +45,16 @@ class CheckYourAnswersController @Inject() (
...
-        Redirect(routes.AmendClaimController.start())
+    data.getAmendAnswers flatMap { answers =>
+      try {
+        val amendClaim = AmendClaim(answers)
+        data.updateAmendAnswers(answers => answers.copy(changePage = None)) map { updatedAnswers =>
+          Ok(checkYourAnswersView(amendClaim, backLink(updatedAnswers)))
+        }
+      } catch {
+        case _: MissingAnswersException =>
+          Future(Redirect(navigator.firstMissingAnswer(answers)))
```
SteveSugden, line 56: We might want to consider a similar "error" CYA page for amend.  But in the case of amend case, the problem should only ever happen if the user deliberately changes the url to `/check-answers`.   There are no answers that become mandatory based on a previous answer.    Saying "yes" to having a file to upload and then not uploading a file is currently acceptable.

crosslandwa:
I'm concerned about the maintenance of `CheckYourAnswersWithMissingView.scala.html` - it looks remarkably similar to the existing `CheckYourAnswersView.scala.html` with the key difference being it takes a `CreateAnswers` instance instead of a `Claim` instance

I see the following maintenance issues
1. duplication of markup (~350 lines) for the summary list (between the two views)
1. duplication of logic about what questions must be answered given answers to previous questions (between the Navigator/Claim and the new view, e.g. where `if(answers.repayTo.contains(RepayTo.Importer))` is used to render the importers EORI)
1. duplication of logic to process the answers into a claim (e.g. the "summing up of the duty types" in both the `Claim` and the `CYAMissingAnswers` view)


I suggest issue 1 above can be mitigated by having a _single_ view with a "signature" like
```
@(claim: Option[Claim], answers: CreateAnswers, missingPage: Page, backLink: NavigatorBack)(implicit request: Request[_], messages: Messages)
```

which internally would use:
 - the `answers` object to "fill in" the SummaryList (as per the proposed `CheckYourAnswersWithMissingView.scala.html`)
 - the absence of the `Option[Claim]` to render the warning at the top of the page
 - the presence of the `Option[Claim]` to decide to render the `Submit` or `Resolve` buttons at the bottom of the page
 
The CYA controller would only pass the `Claim` if it was able to create it from the given `Answers`
 
This offers the advantage of handling CYA with a single view, but it a the cost of the "happy path" rendering of CYA being driven off the `Answers` (rather than the `Claim`, which is what the service will _actually_ be sending to the CMS)

Thoughts?

crosslandwa (approved):

crosslandwa:
I'm happy the extracted `CreateClaimSummary.scala.html` addresses the concerns I outlined above

PR #147, opened 2021-03-26
Nf 334 audit amend

rorticus0502:


app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/services/ClaimService.scala
```diff
@@ -71,6 +77,10 @@ class ClaimService @Inject() (auditConnector: AuditConnector, connector: NIDACCo
     )
 
     connector.amendClaim(AmendEISClaimRequest(eisRequest, amendClaim.uploads), correlationId)
+      .map { response =>
+        auditConnector.sendExplicitAudit("AmendClaim", AmendClaimAudit(response.error.isEmpty, amendClaim, response))
```
crosslandwa, line 81: CLARIFICATION: Will this code be executed in the result of a 500 response from the backend?

test/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/services/ClaimServiceSpec.scala
```diff
@@ -46,6 +47,7 @@ class ClaimServiceSpec extends UnitSpec with BeforeAndAfterEach with TestData {
     super.beforeEach()
 
     when(nidacConnector.submitClaim(any(), any())(any())).thenReturn(Future.successful(validCreateClaimResponse))
+    when(nidacConnector.amendClaim(any(), any())(any())).thenReturn(Future.successful(validAmendClaimResponse))
```
njmoorewitchcat, line 50: OBSERVATION: In regards to the clarification comment above, should there be a tests that do a Future.failure() return here and then that would explain what the audit would do under these circumstances - i.e. if it should still audit if the upstream fails or not.

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/services/ClaimService.scala
rorticus0502: No

rorticus0502: In line with what we discussed with Nathan. If it is rejected we will audit it, but if some unexpected failure downstream, we wont, and that will be covered by implicit audits.

test/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/services/ClaimServiceSpec.scala
rorticus0502: Test for failed future added to document expected behaviour

crosslandwa (approved):

PR #150, opened 2021-03-31
NF-329 - Collect a record of what I just submitted

SteveSugden:


app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/views/makeclaim/ConfirmationView.scala.html
```diff
@@ -15,22 +15,49 @@
...
+@(receipt: CreateClaimReceipt)(implicit request: Request[_], messages: Messages)
+
+@summaryContent = {
+    @claimSummaryHeading(receipt)
+    @claimSummary(receipt.answers, withActions = false)
+}
 
 @layout(pageTitle = InformationTitle(messages("confirmation.title")), NavigatorBack(None)) {
 
+  <div class="nidac-print-hidden">
```
crosslandwa, line 39: Can this be
```suggestion
  <div class="govuk-!-display-none">
```

and the custom styles for `.nidac-print-hidden` removed?

SteveSugden: Can be if it does the same thing.  I'll check.

SteveSugden: It doesn't.   `govuk-!-display-none` never displays anything.  What I want is to be able to mark that block so that it displays normally _except_ when printing (when it's not displayed)

crosslandwa: Sorry, wrong class in my previous comment, i meant `govuk-!-display-none-print`

SteveSugden: That does the trick!   I'm a dope!  It's exactly the same requirement as hiding back and change links when printing - just on a bigger scale!

crosslandwa: Or `nidac-!-display-only-print` to be "inspired" by the govuk style naming

app/assets/stylesheets/media.scss
```diff
@@ -0,0 +1,21 @@
...
+.nidac-!-display-only-print - is used to hide content except when printing
+====================================
+*/
+@media all {
+  .nidac-\21-display-only-print { display: none}
+}
+@media print {
+  @page { margin: 10px }
+  .govuk-section-break { page-break-inside : avoid }
+  .govuk-link { display: none; }
```
crosslandwa, line 15: CLARIFICATION: Is the `govuk-link` styling still needed?

I can see the "nidac-print-link" already has the `govuk-!-display-none-print` class, and the "action links" in the summary table are omitted on the confirmation/review pages by being rendered with `withActions = false`

crosslandwa: Ah - the answer is in the comment - presumably this is needed to hide the sign out link in the header...

SteveSugden: Yes.  I borrowed this file from another project that does printing.   It might be a bit over-kill to hide all links (e.g. there might be links in content that would need to be displayed to make sense when printed).

So it might be better to remove this line as you suggest and to ensure that we explicitly add  `govuk-!-display-none-print` to links that we definitely don't want printed (link "sign-out" and "change" links on regular CYA).

(The `withActions = false` thing was added so that we can render the "summary" on screen without the change links)

crosslandwa (approved):

PR #152, opened 2021-04-07
NF-372 - sign-out url for non-authorised users

SteveSugden:


crosslandwa (approved):

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/views/Layout.scala.html
```diff
@@ -56,7 +56,13 @@
 }
 
 @signOutHref = @{
-    if(logout) Some(SignOutController.signOut().url) else None
+    if(logout) {
```
crosslandwa, line 59: OBSERVATION: I wonder if there's a way to use `match` to neaten up the nested `if|else` here...

SteveSugden: Could do.  For the lols

test/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/views/ServiceUnavailablePageViewSpec.scala
```diff
@@ -38,7 +40,7 @@ class ServiceUnavailablePageViewSpec extends UnitViewSpec {
     }
 
     "have sign out link" in {
```
crosslandwa, line 42: Perhaps capture the reason for this sign out link being "different" in the test name, maybe something like

```suggestion
    "have sign out link (for users that cannot access the NIDAC service)" in {
```

crosslandwa (approved):

PR #153, opened 2021-04-07
NF-371 Zap fix, now escape the HtmlContents that are merged data

njmoorewitchcat:


app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/views/makeclaim/CreateClaimSummary.scala.html
```diff
@@ -99,7 +100,7 @@
                     messages("check_answers.claim.entryDetails.epu", entryDetails.entryProcessingUnit),
                     messages("check_answers.claim.entryDetails.entryNumber", entryDetails.entryNumber),
                     messages("check_answers.claim.entryDetails.entryDate", DateFormatter.format(entryDetails.entryDate))
-                ).mkString("<br>")).getOrElse(""))
+                ).map(HtmlFormat.escape(_)).mkString("<br>")).getOrElse(""))
```
njmoorewitchcat, line 103: OBSERVATION: This is not necessary in this block as they are all generated from messages and not directly used from user entry fields.  Will remove if someone else doesn't like it here.

rorticus0502 (approved):

PR #154, opened 2021-04-09
NF-386 send correct due amount to EIS

rorticus0502:


njmoorewitchcat (approved):

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/models/eis/DutyDetail.scala
```diff
@@ -30,7 +30,7 @@ object DutyDetail {
     new DutyDetail(
       reclaiming,
       BigDecimal.apply(paid.actuallyPaid).setScale(decimalPlaces).toString(),
-      paid.dueAmount.setScale(decimalPlaces).toString
+      BigDecimal.apply(paid.shouldHavePaid).setScale(decimalPlaces).toString
     )
```
SteveSugden, line 34: Can you rename that "dueAmount" val to something like "returnAmount"

PR #155, opened 2021-04-09
NF-374 - multi-file upload on single page

SteveSugden:


app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/views/Head.scala.html
```diff
@@ -30,6 +30,8 @@
     <style type="text/css">
             a[href^="javascript:"].govuk-back-link { visibility: hidden; }
             a[href^="javascript:"].nidac-print-link { display: none; }
+            div.nidac-display-only-non-javascript { display: contents; }
```
rorticus0502, line 33: I think this was mentioned in one of the three amigo sessions, but I cant remember.
Why do we have css here in the html, but also have similar css in seperate css file upload-loading.scss ?

njmoorewitchcat (approved):
LGTM

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/views/Head.scala.html
SteveSugden: The one in the .scss file "turns off" the display - that is the default.   Here we are turning it on for non-javascript users.  So `nidac-display-only-non-javascript` is a class to use for elements that should _only_ be seen when JS is disabled.

rorticus0502 (approved):

PR #158, opened 2021-04-12
NF-363 Preserve logging context around Mongo calls

crosslandwa:
Fixes bug where tracing context (eg. x-request-id) would be lost from logs for any code executing within the context of mongo query


SteveSugden (approved):
Can I be a pain and suggest a comment in both files about why this unusual feature is required.

SteveSugden (dismissed):

SteveSugden (dismissed):

SteveSugden (approved):

PR #160, opened 2021-04-13
NF-374 - bug fix for non-javascript users

SteveSugden:
UPSCAN and UPSCAN_STUB return different error codes for "missing file"

### DIFF
--- a/app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/controllers/FileUploading.scala
+++ b/app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/controllers/FileUploading.scala
@@ -30,9 +30,12 @@ import scala.concurrent.{ExecutionContext, Future}
 
 trait FileUploading {
 
-  protected val DUPLICATE    = "DUPLICATE"
-  protected val UNKNOWN      = "UNKNOWN"
-  protected val MISSING_FILE = "InvalidArgument"
+  protected val ERROR_DUPLICATE      = "DUPLICATE"
+  protected val UNKNOWN              = "UNKNOWN"
+  protected val ERROR_MISSING_FILE   = "InvalidArgument"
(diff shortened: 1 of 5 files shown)


app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/controllers/FileUploading.scala
```diff
@@ -64,17 +67,19 @@ trait FileUploading {
...
-      case "400" | MISSING_FILE => error("error.file-upload.required")
-      case "InternalError"      => error("error.file-upload.try-again")
-      case "EntityTooLarge"     => error("error.file-upload.invalid-size-large")
-      case "EntityTooSmall"     => error("error.file-upload.invalid-size-small")
-      case "QUARANTINE"         => error("error.file-upload.quarantine")
-      case "REJECTED"           => error("error.file-upload.invalid-type")
-      case DUPLICATE            => error("error.file-upload.duplicate")
-      case _                    => error("error.file-upload.unknown")
+      case "400" | ERROR_MISSING_FILE => error("error.file-upload.required")
+      case "InternalError"            => error("error.file-upload.try-again")
```
njmoorewitchcat, line 76: CLARIFICATION: Should we make all the strings here constants or is there a reason not too?

crosslandwa (approved):

njmoorewitchcat (approved):

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/controllers/FileUploading.scala
SteveSugden: I just wanted to "expose" certain codes.  But there's no reason not to do as you suggest.  Next time, maybe.

PR #172, opened 2021-04-21
NF-410 NF-411 NF-422 content changes to questions

rorticus0502:
content changes to questions based on RAD

### DIFF
--- a/conf/messages
+++ b/conf/messages
@@ -30,7 +30,7 @@ service.unavailable.paragraph.1 = We are working on updating this service.
 service.unavailable.paragraph.2 = In the meantime you can continue to submit NIDAC claims using the existing method.
 
 representation_type.title = Are you the importer?
-representation_type.error.required = Select if you are the importer or their representative
+representation_type.error.required = Confirm if you are the importer or their representative
 representation_type.Importer = Yes
 representation_type.Representative_of_importer = No, I am a representative of the importer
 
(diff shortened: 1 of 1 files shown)


conf/messages
```diff
@@ -29,10 +29,10 @@ service.unavailable.title = Sorry, the service is unavailable
 service.unavailable.paragraph.1 = We are working on updating this service.
 service.unavailable.paragraph.2 = In the meantime you can continue to submit NIDAC claims using the existing method.
 
-representation_type.title = Are you the importer or their representative?
-representation_type.error.required = Select if you are the importer or their representative
-representation_type.Importer = I am the importer
-representation_type.Representative_of_importer = I am a representative of the importer
+representation_type.title = Are you the importer?
+representation_type.error.required = Confirm if you are the importer or their representative
```
SteveSugden, line 33: OBSERVATION:  That should really be something like "Select yes if you are the importer" to conform to guidelines

SteveSugden (approved):

conf/messages
rorticus0502: I asked for guidance from design and this this what they came up with

rorticus0502: I prefer your suggestion actually, it is more instructive

SteveSugden: Thanks.   FYI - https://design-system.service.gov.uk/components/radios/#error-messages

PR #182, opened 2021-04-23
NF-113 - upgrade to Play 2.8 and other dependency updates

SteveSugden:


app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/models/create/CreateAnswers.scala
```diff
@@ -62,5 +62,8 @@ final case class CreateAnswers(
 
 object CreateAnswers {
 
+  // import required since play 2.8 to discover implicit read/writes of Enumerable
```
rorticus0502, line 65: can we remove the comment

test/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/views/makeclaim/BusinessNameViewSpec.scala
```diff
@@ -19,10 +19,7 @@ package uk.gov.hmrc.nationalimportdutyadjustmentcentrefrontend.views.makeclaim
 import org.jsoup.nodes.Document
 import play.api.data.Form
 import uk.gov.hmrc.nationalimportdutyadjustmentcentrefrontend.base.{TestData, UnitViewSpec}
-import uk.gov.hmrc.nationalimportdutyadjustmentcentrefrontend.forms.create.{
```
rorticus0502, line 22: thank you

rorticus0502 (approved):

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/models/create/CreateAnswers.scala
SteveSugden: IntelliJ shows the import as unused.  Without the import, the error message on compilation is quite cryptic and it took me quite a while to work out this "fix".  

So on reflection, I thought the comment was appropriate in case our further selves are tempted to remove the import.

BTW - Other "fixes" might be possible. Play 2.8 introduces Jackson serialisation support, so it might be possible to change the way we define our implicit read/writes - but this would need further investigation.

PR #189, opened 2021-04-28
NF-458 ALF url and label config

rorticus0502:


app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/models/addresslookup/AddressLookupRequest.scala
```diff
@@ -36,23 +36,45 @@ case class AddressLookupRequest(
 object AddressLookupRequest {
   implicit val format: OFormat[AddressLookupRequest] = Json.format[AddressLookupRequest]
 
-  def apply(continueUrl: String, lookupPageHeadingKey: String)(implicit
+  def apply(continueUrl: String, homeUrl: String, signOutUrl: String, lookupPageHeadingKey: String, hintKey: String)(
+    implicit
     messagesApi: MessagesApi,
     config: AppConfig
```
SteveSugden, line 42: OBS:  This method already has the appConfig and yet we're passing in values from appConfig.  Could it not do one or the other?

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/config/AppConfig.scala
```diff
@@ -91,6 +91,8 @@ class AppConfig @Inject() (
   val importerAddressLookupCallbackUrl: String =
     s"$loginContinueUrl/create${controllers.makeclaim.routes.ImporterDetailsController.onUpdate("").url}"
 
+  val addressLookupShowPhaseBanner = true
```
SteveSugden, line 94: Should this be derived from application.conf?   And as a possible enhancement, we could use this flag to drive the setting of the phase banner in our own service (in Layout.scala.html) - then they can be turned off in sync.

SteveSugden (approved):
LGTM

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/models/addresslookup/AddressLookupRequest.scala
njmoorewitchcat: Continue URL and lookupPage Heading and possibly hintKey are controller specific (so different between your address and importer address), not sure on the others bit I think if they are generic they could be drawn straight from the config object.

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/config/AppConfig.scala
njmoorewitchcat: +1

njmoorewitchcat (approved):

PR #193, opened 2021-04-29
NF-459 Fix Keepalive Url addition slash

njmoorewitchcat:


SteveSugden (approved):
I'll approve for now.  But I think we should re-visit the mis-use of the "login" url base.

PR #194, opened 2021-04-29
BAU - proposed change to address lookup url config

SteveSugden:


rorticus0502 (dismissed):

njmoorewitchcat (approved):
Took me a while, but I get it - pass the path in to the connector then attach the selfBaseURL.  LGTM - making the campsite tidier.

rorticus0502 (approved):

PR #197, opened 2021-04-30
BAU - better handling when upscan record not found

SteveSugden:


app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/repositories/UploadRepository.scala
```diff
@@ -113,12 +113,12 @@ class UploadRepository @Inject() (mongoComponent: MongoComponent, config: AppCon
...
 
-  def updateStatus(reference: Reference, journeyId: JourneyId, newStatus: UploadStatus): Future[UploadStatus] =
+  def updateStatus(reference: Reference, journeyId: JourneyId, newStatus: UploadStatus): Future[Boolean] =
     Mdc.preservingMdc {
-      collection.findOneAndUpdate(
+      collection.updateOne(
         and(equal("reference", Codecs.toBson(reference)), equal("journeyId", Codecs.toBson(journeyId))),
         set("status", Codecs.toBson(newStatus))
-      ).toFuture() map (details => details.status)
+      ).toFuture() map (result => result.getModifiedCount == 1)
```
rorticus0502, line 121: I prefer this, returning a boolean clearly indicating if exactly what we intended, did indeed happen.

rorticus0502 (approved):

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/repositories/UploadRepository.scala
SteveSugden: Yes.  Returning the same status we passed in was a bit pointless.   BTW this is where the null pointer was happening - details is null when the record is not found with "findAndUpdateOne".   "updateOne" just returns what happened.

crosslandwa (approved):

PR #198, opened 2021-04-30
NF-469 moved address change link in line with Continue button and add…

rorticus0502:
…ed some css to space it

### DIFF
--- a/app/assets/stylesheets/upload-loading.scss
+++ b/app/assets/stylesheets/upload-loading.scss
@@ -34,3 +34,7 @@
 .nidac-display-only-non-javascript {
   display: none;
 }
+
+.address-change-link {
+    padding-left: 15px;
+}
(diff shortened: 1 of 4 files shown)


app/assets/stylesheets/upload-loading.scss
```diff
@@ -34,3 +34,7 @@
 .nidac-display-only-non-javascript {
   display: none;
 }
+
+.address-change-link {
+    padding-left: 15px;
```
SteveSugden, line 39: This file is becoming a "dumping ground" for non upload related stuff.  Anyway...

Are you aware of these built-in styles?  `govuk-!-padding-left-x`. `govuk-!-padding-right-x`
They only go 0-9 so you might need to have a left (link) and right (button)

SteveSugden (approved):
Lovely

crosslandwa (approved):

PR #210, opened 2021-05-11
NF-480 - duty payments 2 d.p.

SteveSugden:


app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/models/create/DutyPaid.scala
```diff
@@ -23,5 +23,12 @@ case class DutyPaid(actuallyPaid: String, shouldHavePaid: String) {
 }
 
 object DutyPaid {
+  private val decimalPlaces              = 2
   implicit val format: OFormat[DutyPaid] = Json.format[DutyPaid]
+
+  private def setScale(amount: String) = BigDecimal.apply(amount).setScale(decimalPlaces).toString()
```
crosslandwa, line 29: naming this function `def to2DP` would make its intent clearer IMO

crosslandwa: I would agree with you if the number of decimal places could be specified by the calling code (i.e. passed in as a parameter). As its written, `setScale` will **always** return a 2dp value - the fact that the `decimalPlaces` constant has been extracted is an implementation detail that clients of `setScale` can't influence/shouldn't care about

SteveSugden: You are right.  Client's shouldn't care what scale is set.  As longs as it's set.   Imagine if the definition of `decimalPlaces` was changed to 3.  Then the client caller of `to2dp` would be confused.  The caller of `setScale` would continue not to care.

crosslandwa: > Client's shouldn't care what scale is set

I disagree - clients that are using this function to format a value for display to users _do care_ how many DP the response has - if it started returning something different, e.g. 3DP, it _may_ longer be suitable for their needs. What clients shouldn't care about is how the that function is implemented internally

Named as it is, `setScale`
 - doesn't tell the calling code what it will get back
 - suggests it may update ("set") the state of something
 - demands that the developer writing code that calls `setValue` read its implementation to see what it does
 
That said, given this is a small private function this is all somewhat moot as the only clients of this one liner are in the same file...

PR #211, opened 2021-05-17
NF-275 - the "upload in progress page" refreshes automatically every 1 sec

SteveSugden:


rorticus0502 (approved):
Used the file upload and it works as advertised.
Code looks good

PR #212, opened 2021-05-18
NF-507 - encrypt user answers in Mongo

SteveSugden:


app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/models/CacheData.scala
```diff
@@ -18,28 +18,75 @@ package uk.gov.hmrc.nationalimportdutyadjustmentcentrefrontend.models
...
+
+object ProtectedData {
+  private val formats: OFormat[ProtectedData] = Json.format[ProtectedData]
+  private val jsonEncryptor                   = new JsonEncryptor()(CryptoProvider.crypto, formats)
+  private val jsonDecryptor                   = new JsonDecryptor()(CryptoProvider.crypto, formats)
+
+  implicit val reads: Reads[ProtectedData] = Reads {
+    json =>
+      if (CryptoProvider.enabled)
+        Json.fromJson[Protected[ProtectedData]](json)(jsonDecryptor).map(_.decryptedValue)
```
crosslandwa, line 79: OBSERVATION: I'd consider moving the encryption/decryption code into `CryptoProvider` and exposing `encrypt(input): String` and `decrypt(encrypted: String)` methods (to collect all "encryption" concerns in that class)

Internally, the encrypt/decrypt methods could make use of the `enabled` property to return the input "as is, but as a serialized string" (if disabled) or as an encrypted string (if enabled)

SteveSugden: I'm not sure it would work with the encrypt/decrypt methods you mention.  But it might be possible to create some generic read/writes methods on the "Provider".  I'll take a look.

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/models/CacheData.scala
```diff
@@ -18,28 +18,75 @@ package uk.gov.hmrc.nationalimportdutyadjustmentcentrefrontend.models
...
-  amendAnswers: Option[AmendAnswers] = None,
-  createClaimReceipt: Option[CreateClaimReceipt] = None,
-  amendClaimReceipt: Option[AmendClaimReceipt] = None,
+  data: ProtectedData = ProtectedData(),
   lastUpdated: Instant = Instant.now()
 ) {
 
-  def getCreateAnswers: CreateAnswers = createAnswers.getOrElse(CreateAnswers())
-  def getAmendAnswers: AmendAnswers   = amendAnswers.getOrElse(AmendAnswers())
+  val createAnswers      = data.createAnswers
```
njmoorewitchcat, line 36: AMBER: The top two are redundant.  Should we remove all four of these and refactor the usages to draw them out of the inner data object instead?

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/models/CacheData.scala
crosslandwa: i.e. this code would look something like

```
implicit val reads: Reads[ProtectedData] = Reads {
  json => Json.fromJson(CryptoProvider.decrypt(json))
}

implicit val writes: Writes[ProtectedData] = Writes { data =>
  CryptoProvider.encrypt(Json.toJson(data)(formats))
}
 ```
and the `jsonEncryptor` and `jsonDecryptor` fields would be encapsulated within `CryptoProvider`

File: app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/models/CacheData.scala (hunk 2 above)
Comment (SteveSugden) [reply app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/models/CacheData.scala:36]: They are not redundant.  They replace the removed fields from the constructor and were added to minimise the impact on the rest of the code-base.

File: conf/application.conf [hunk 3]
```diff
@@ -194,4 +194,4 @@ eori {
 
 phaseBanner.display = true
 
-
+json.encryption.key = "P5xsJ9Nt+quxGZzB4DeLfw=="
```
crosslandwa, line 197: CLARIFICATION: Have I understood correctly that to enable encryption would we need _two_ pieces of config present:
- `json.encryption.key`
- `json.encryption.enabled`

?

If so, whats the reasoning for driving this from two pieces of config, over using the presence of `json.encryption.key` to enable/disable encryption?

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/models/CacheData.scala
crosslandwa: ^ With the above suggested approach the `ProtectedData` class is responsible for doing serialisation to/from JSON strings, and the `CryptoProvider` is encrypt/decrypting those (JSON) strings (so no generics)

conf/application.conf
njmoorewitchcat: You would not want the encryption to be off, if you forgot to add an encryption key in a higher environment.  You don't want to publish without an encryption key.  So it looks like we put a default key in and you can only switch it off if you explicitly set the boolean to false.  If the key is missing, you get an error.  If the boolean is missing, then it defaults to true.  Therefore, in forgetful scenarios, you get default encryption switch on.

SteveSugden: The second config value (which is missing from default config) is just there to disable encryption.  I suppose if could be triggered by the absence of the encryption key but that would be an unusual way of turning something off

njmoorewitchcat: TLDR:  The absence of the key should cause an error - not default to no encryption!

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/models/CacheData.scala
crosslandwa: I'd spotted that we have _two ways_ to access some of the inner data here - I think this is an existing issue and should be addressed separately (as the ripple out effect of unifying to a single access method could be significant...)

conf/application.conf
SteveSugden: I agree.  That's the way it is written

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/models/CacheData.scala
SteveSugden: Agreed.  I think the users of CacheData should be isolated from this change as much as possible

conf/application.conf
crosslandwa: Makes sense - agree we should have **_encryption enabled by default_** and either:
 - blow up if `json.encryption.key` is missing
 - only disable encryption if our config contains `json.encryption.enabled: false`
 
I've just re-read the boolean logic in `CryptoProvider` and have confirmed that I think it will give us "encryption enabled by default" 👍

crosslandwa: Is it worth updating the `README: Local Setup` documentation to describe this (and how to turn off encryption locally if desired)?

SteveSugden: I only added the disable switch to keep Rory happy!  Now I've created more work.

crosslandwa: I think its a valuable change/option

I'd misinterpreted the logic/config needed to enable/disable the encryption - will resolve this conversation!

rorticus0502: I thought we agreed to have the ability to disable encryption locally in a seperate ticket, but thank you still.
Seriously though whilst it may be useful to disable it locally it is not required as part of this ticket IMO

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/models/CacheData.scala
njmoorewitchcat: I have pulled the branch and removed both `val createAnswers` and `val amendAnswers` , and it compiled and run the unit tests (didn't run the IT's).  Where may they be used? I suspect we use the getCreateAnswers and getAmendAnswers instead and therefore don't need these two public vals.

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/models/CacheData.scala
SteveSugden: After much refactoring I eventually ended up with methods as you suggested.

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/models/CacheData.scala
SteveSugden: You are right.  I added the 2 new vals before checking that they were actually needed anywhere - they weren't.

njmoorewitchcat (approved):

crosslandwa (approved):

PR #216, opened 2021-05-20
NF-512 - allow zero value for "should have paid"

SteveSugden:


test/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/forms/create/DutyPaidFormProviderSpec.scala
```diff
@@ -0,0 +1,77 @@
...
+    behave like mandatoryField(form, fieldName, requiredError = FormError(fieldName, requiredKey))
+
+    behave like fieldThatPreventsUnsafeInput(
+      form,
+      fieldName,
+      nonNumerics,
+      FormError(fieldName, invalidKey, Seq(Validation.dutyPattern))
+    )
+
+    "error when input is negative" in {
```
njmoorewitchcat, line 70: OBSERVATION: As this is a new test around a boundary condition, I would have liked to see tests that cover that.  So a zero test that passes, a -0.01 test that fails.  Maybe also a -0.000000001 that fails and a 0.00 that passes (as these are string fields)

njmoorewitchcat (approved):

njmoorewitchcat (approved):

PR #217, opened 2021-05-21
NF-507 - refactor to fix restart issue

SteveSugden:
Changing CryptoProvider to a class fixes the issue where ConfigFactory fails to load
application.conf after sbt recompiles/restarts (Classloader problem)


rorticus0502:
> Changing CryptoProvider to a class fixes the issue where ConfigFactory fails to load
> application.conf after sbt recompiles/restarts (Classloader problem)

Nice

rorticus0502 (approved):

njmoorewitchcat (approved):

PR #222, opened 2021-05-27
BAU - remove "page name" map from navigators

SteveSugden:


app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/pages/Page.scala
```diff
@@ -23,3 +23,8 @@ case object RequiredDocumentsPage extends Page
 case object UploadPage            extends Page
 case object CheckYourAnswersPage  extends Page
 case object ConfirmationPage      extends Page
+
+object Implicit {
```
crosslandwa, line 27: CLARIFICATION: I'd appreciate a quick talk-through of how this works (it looks slightly voodoo-ish to my naive eyes). I assume type-checking is still provided in places that import `pages.Implicit._` then reference "pages"?

test/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/views/amendclaim/CheckYourAnswersViewSpec.scala
```diff
@@ -90,7 +89,7 @@ class CheckYourAnswersViewSpec extends UnitViewSpec with TestData {
           s"${messages("site.change")} ${messages("amend.check_answers.information.uploadedDocuments.accessible")}"
         )
 
-        uploadRow must haveSummaryActionsHref(routes.CheckYourAnswersController.onChange(AmendPageNames.uploadSummary))
+        uploadRow must haveSummaryActionsHref(routes.CheckYourAnswersController.onChange(UploadPage))
```
crosslandwa, line 92: CLARIFICATION: How does one distinguish if `UploadPage` refers to the create claim upload page or the amend claim upload page? Does it matter?

crosslandwa (approved):

app/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/pages/Page.scala
SteveSugden: Yes.  We can have a chat later.   But places that require a `Page` will still enforce that a `Page` is provided.    This implicit type conversion is to allow a `Page` to be passed to a method that requires a `String` (without having to type `SomePage.toString`)

test/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/views/amendclaim/CheckYourAnswersViewSpec.scala
SteveSugden: It doesn't matter.  As long as that `Page` only occurs once in the `pageOrder: Seq[P]` definition

crosslandwa (approved):

PR #229, opened 2021-06-11
BUG - don't include id parameter in callback url to ALF

SteveSugden:


crosslandwa (dismissed):
OBSERVATION: It _may_ be helpful to add something to the controller specs to verify the `callbackURL` passed to the `AddressLookupService` is _as we'd expect_ (i.e. without the trailing `?id=`) - i'm thinking as a way of documenting intent (rather than to verify the _correctness_ of the URL, as we'll find that out when we try and integrate with ALF in QA)

rorticus0502 (approved):

PR #232, opened 2021-06-18
NF-591 - ALF confirmation screen headings

SteveSugden:


test/uk/gov/hmrc/nationalimportdutyadjustmentcentrefrontend/models/addresslookup/AddressLookupRequestSpec.scala
```diff
@@ -98,4 +85,21 @@ class AddressLookupRequestSpec extends UnitSpec with MockitoSugar with Injector
     )
   )
 
+  "AddressLookupRequest" should {
```
njmoorewitchcat, line 88: AMBER: We need a second test here that passes the "importer-details.confirmation.heading" in (and "importer-details.hint" .  As the initial references are just strings, without a second test, we currently could delete the "importer-details.confirmation.heading" key/value from the message file and no tests would fail.

SteveSugden: That is true of lots of message strings.   We don't have 100% coverage to make sure all expected message keys are actually defined.

njmoorewitchcat (approved):

